PR #58, opened 2018-09-04
PoC: Use tsutils to scan for comments

eps1lon:
Changed the advanced example as @Gerrit0 suggested which fails with the current findComments walker. Tsutils does handle this case however.

This does not include the nodes to the comments.

I tried `Symbol#getDocumentationComment` but this only returned text and stripped text, tags etc.

Adresses #56 

/cc @pgonzal

### DIFF
--- a/api-demo/assets/advanced-input.ts
+++ b/api-demo/assets/advanced-input.ts
@@ -1,3 +1,4 @@
+/** */
 export class Statistics {
   /**
    * Returns the average of two numbers.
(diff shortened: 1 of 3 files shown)


Gerrit0:
I don't think adding another dependency should be required, though it might be worth looking at the source of [`forEachComment`](https://github.com/ajafff/tsutils/blob/4099ae6aea71270c97e618092de31936da2f4a03/util/util.ts#L453) and extracting it's functionality.

eps1lon:
Succeeded by #59

PR #94, opened 2018-10-09
Remove empty DIVs by using React.Fragment & paddingTop

KevinTCoughlin:
Removed some empty DIVs in favor of [React.Fragment](https://reactjs.org/docs/fragments.html) from React@16 and use of `padding-top`. Hope you don't mind the nit changes, just getting acquainted with the source.

## Changes

- Removes empty render DIVs with React.Fragment since fragments do not add to node count.
- Replaces empty DIV of `height: 40px` in favor of parent element `padding-top: 40px`.

## Test

- Monaco Editor container top should render in-line with top of tab content container

**After**

image

- Empty DIVs should be removed, total node count reduced by 3

**Before**

```js
document.getElementsByTagName('*').length
202
```

**After**

```js
document.getElementsByTagName('*').length
199
```

Minor feature proposals coming soon :)

### DIFF
--- a/playground/src/App.tsx
+++ b/playground/src/App.tsx
@@ -5,11 +5,11 @@ class App extends React.Component {
   public render(): React.ReactNode {
 
     return (
-      <div>
+      <>
         <h1>TSDoc Playground</h1>
 
         <PlaygroundView />
-      </div>
+      </>
     );
   }
 }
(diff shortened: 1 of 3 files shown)


iclanton:
I didn't know you could do `<>...</>`. Neat.

Approved

PR #204, opened 2019-12-19
Add support for synonyms and the 'see' tag

### LINKED ISSUE #203: Support JSDoc "synonyms" for tags
JSDoc supports synonyms for many tags, and it would be worthwhile to support those synonyms for code bases that were previously migrated from JavaScript to TypeScript without requiring an investment in time to rewrite doc comment tags to support TSDoc. For example:

- `@abstract` (synonyms: `@virtual`)
- `@augments` (synonyms: `@extends`)
- `@default` (synonyms: `@defaultvalue`)
- `@description` (synonyms: `@desc`)
- `@fires` (synonyms: `@emits`)
- `@param` (synonyms: `@arg`, `@argument`)
- `@returns` (synonyms: `@return`)
- `@throws` (synonyms: `@exception`)
- `@yields` (synonyms: `@yield`)
- `{@link}` (synonyms: `{@linkcode}`, `{@linkplain}`)
- etc.

It would also be valuable for type parameter support (#72), allowing `@template` as a synonym for `@typeParam`.

rbuckton:
This adds support for synonyms for tags, and adds the `@see` tag from JSDoc.

Fixes #203


rbuckton:
@octogonz Can you take a look at this and tell me what you think?

avin-kavish:
hello from 2022!

PR #334, opened 2022-08-26
Fixed Sev2 accessibility bugs for TSDoc Playground

Hashah1:
Added fixes for the playground view when zoomed in to 400%. See https://github.com/microsoft/tsdoc.org-website/issues/35

### DIFF
--- a/playground/package.json
+++ b/playground/package.json
@@ -15,8 +15,7 @@
     "react": "18.2.0",
     "monaco-editor": "~0.17.1",
     "promise": "~8.0.3",
-    "tslib": "~1.10.0",
-    "react-media": "~1.10.0"
+    "tslib": "~1.10.0"
   },
   "devDependencies": {
     "@rushstack/eslint-config": "~2.3.1",
--- a/playground/package.json
+++ b/playground/package.json
@@ -12,7 +12,6 @@
     "@microsoft/tsdoc": "0.14.1",
     "@types/react-dom": "16.9.3",
     "@types/react": "16.9.11",
-    "react": "18.2.0",
     "monaco-editor": "~0.17.1",
     "promise": "~8.0.3",
     "tslib": "~1.10.0"
--- a/playground/package.json
+++ b/playground/package.json
@@ -12,9 +12,11 @@
     "@microsoft/tsdoc": "0.14.1",
     "@types/react-dom": "16.9.3",
     "@types/react": "16.9.11",
+    "react": "18.2.0",
     "monaco-editor": "~0.17.1",
     "promise": "~8.0.3",
-    "tslib": "~1.10.0"
+    "tslib": "~1.10.0",
+    "react-media": "~1.10.0"
   },
   "devDependencies": {
     "@rushstack/eslint-config": "~2.3.1",
--- a/playground/src/PlaygroundView.css
+++ b/playground/src/PlaygroundView.css
@@ -1,0 +1,12 @@
+.playground-main-row {
+  align-items: stretch;
+  flex: 1;
+}
+
+@media screen and (max-width: 700px) {
+  .playground-main-row {
+    align-items: stretch;
+    flex: 1;
+    flex-direction: column !important;
+  }
+}
--- a/playground/src/PlaygroundView.tsx
+++ b/playground/src/PlaygroundView.tsx
@@ -1,5 +1,7 @@
 import * as React from 'react';
 import * as tsdoc from '@microsoft/tsdoc';
+
+// TODO: Fix this to be complied as .module.scss
 import './PlaygroundView.module.scss';
 
 import { TabPane } from './TabPane';
--- a/playground/src/PlaygroundView.tsx
+++ b/playground/src/PlaygroundView.tsx
@@ -1,5 +1,8 @@
+/* eslint-disable @typescript-eslint/explicit-member-accessibility */
+/* eslint-disable @typescript-eslint/member-ordering */
 import * as React from 'react';
 import * as tsdoc from '@microsoft/tsdoc';
+import Media from 'react-media';
 
 import { TabPane } from './TabPane';
 import { FlexRowDiv, FlexColDiv } from './FlexDivs';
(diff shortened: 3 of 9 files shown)


playground/src/PlaygroundView.tsx
```diff
@@ -1,5 +1,8 @@
+/* eslint-disable @typescript-eslint/explicit-member-accessibility */
+/* eslint-disable @typescript-eslint/member-ordering */
```
iclanton, line 2: Can you fix these?

playground/src/PlaygroundView.tsx
```diff
@@ -66,22 +69,41 @@ export class PlaygroundView extends React.Component<IPlaygroundViewProps, IPlayg
...
+  };
+  mainRowStyleZoomed: React.CSSProperties = {
+    alignItems: 'stretch',
+    flex: 1,
+    flexDirection: 'column'
+  };
+  errorsPaneStyle: React.CSSProperties = {
+    height: '130px',
+    marginTop: '20px'
+  };
```
iclanton, line 84: These can go into consts outside the class.

playground/src/PlaygroundView.tsx
```diff
@@ -67,21 +82,26 @@ export class PlaygroundView extends React.Component<IPlaygroundViewProps, IPlayg
...
+        <Media query="(min-width: 700px)">
+          {
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            (matches: any) =>
+              matches
+                ? this._renderLargeScreenPlayground() // For example: Laptops
+                : this._renderSmallScreenPlayground() // For example: iPhones/iPads
+          }
+        </Media>
+      </div>
```
samavati, line 96: Why didn't you use simple CSS media queries?

Hashah1: Hello, thank you for the feedback.
I guess more of a personal preference, since I used media query like this in past projects. Would you like me to switch away from this?

samavati: It is more efficient to do such things with CSS than with JS (IMO)

Hashah1: Awesome, made changes to use CSS rather than JS. Thank you!

playground/package.json
```diff
@@ -12,6 +12,7 @@
     "@microsoft/tsdoc": "0.14.1",
     "@types/react-dom": "16.9.3",
     "@types/react": "16.9.11",
+    "react": "18.2.0",
```
iclanton, line 15:
```suggestion
```

React should come from CloudFlare.

This is also a different major version than the `@types` package.

playground/src/PlaygroundView.css
```diff
@@ -0,0 +1,12 @@
+.playground-main-row {
```
iclanton, line 1: We usually use `.module.scss` files, with class names that contain a hash.

Can you convert this file to `.module.scss` and change these class names to something that can be a TS variable name (like `playgroundMainRow`)?

playground/src/PlaygroundView.tsx
```diff
@@ -24,6 +25,11 @@ export interface IPlaygroundViewState {
   selectedTheme: string;
 }
 
+const errorsPaneStyle: React.CSSProperties = {
```
iclanton, line 28:
```suggestion
const ERRORS_PANE_STYLE: React.CSSProperties = {
```

playground/webpack.shared.config.js
```diff
@@ -23,6 +23,10 @@ const MONACO_URL = {
   dev: 'https://cdnjs.cloudflare.com/ajax/libs/monaco-editor/0.14.3/min/',
   production: 'https://cdnjs.cloudflare.com/ajax/libs/monaco-editor/0.14.3/min/'
 };
+const REACT_MEDIA_URL = {
```
iclanton, line 26: You need to list `react-media` as an external (line 89) and include include these URLs in the HTML template (line 109).

Hashah1: `react-media` isnt being used since we switched to simple CSS media queries. Should I delete `REACT_MEDIA_URL`?

iclanton: If it's not actually being used, then yeah

playground/src/PlaygroundView.css
Hashah1: So name the file soemthing like `playgroundView.module.scss`? I looked on the odsp web repo and there were examples of `.module.scss` being used like that

iclanton: Yeah, exactly.

playground/src/PlaygroundView.tsx
```diff
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import * as tsdoc from '@microsoft/tsdoc';
+import './PlaygroundView.module.scss';
```
iclanton, line 3:
```suggestion
import styles from './PlaygroundView.module.scss';
```

playground/src/PlaygroundView.tsx
```diff
@@ -67,21 +73,10 @@ export class PlaygroundView extends React.Component<IPlaygroundViewProps, IPlayg
...
-
-    const errorsPaneStyle: React.CSSProperties = {
-      height: '130px',
-      marginTop: '20px'
-    };
-
     return (
       <FlexColDiv className="playground-content-area" style={{ margin: '4px', flex: 1 }}>
-        <FlexRowDiv className="playground-main-row" style={mainRowStyle}>
+        <FlexRowDiv className="playgroundMainRow">
```
iclanton, line 78:
```suggestion
        <FlexRowDiv className={styles.playgroundMainRow}>
```

playground/src/PlaygroundView.tsx
Hashah1: I made these changes but for some reason the styling isnt applied if I set `className={styles.playgroundMainRow}`. Should I leave as is?

iclanton: I just left a note about needing to fix this later.

iclanton (approved):

PR #375, opened 2024-04-10
Add support for `@since` and `@version` tags

Ratibus11:
Adding a support for `@since` and `@version` tags.

Refers to #136.


Ratibus11:
@microsoft-github-policy-service agree

PR #471, opened 2026-03-31
🚀 Feature: Create DocHeading node type for Markdown headings

### LINKED ISSUE #197: Markdown headings support
Support for markdown headings is listed as supported in the README, but unsupported in https://github.com/microsoft/tsdoc/issues/169. In practice (using the tsdoc playground) it seems to be unsupported. In #169, it is also recommended to open a new issue for further discussion on this point if needed.

To start of the discussion, I will share this example that currently works well with TypeDoc:

```ts
/**
 * Reverse an array in-place.
 *
 * @param array - The mutable array-like object of interest.
 *
 * @param start - The index of the first element in the range to be
 *   reversed, inclusive. The default value is `0`. Negative values
 *   are taken as an offset from the end of the array.
 *
 * @param stop - The index of the last element in the range to be
 *   reversed, inclusive. The default value is `-1`. Negative values
 *   are taken as an offset from the end of the array.
 *
 * #### Complexity
 * Linear.
 *
 * #### Undefined Behavior
 * A `start` or  `stop` index which is non-integral.
 *
 * #### Example
 * ```typescript
 * import { ArrayExt } from '@lumino/algorithm';
 *
 * let data = [0, 1, 2, 3, 4];
 * ArrayExt.reverse(data, 1, 3);  // [0, 3, 2, 1, 4]
 * ArrayExt.reverse(data, 3);     // [0, 3, 2, 4, 1]
 * ArrayExt.reverse(data);        // [1, 4, 2, 3, 0]
 * ```
 */
```

In this comment, the `#### Example` heading can and should be replaced with `@example`. It is not so clear what to do with the `#### Complexity` and `#### Undefined Behavior` headings. A logical first step is to put both in a `@remarks` section, but it is unclear if there is any way to subdivide the remarks section further (without falling back to HTML), other than to allow markdown headers.

lukebaze:
## Problem

Create a new DocNode type to represent Markdown headings in the TSDoc AST. This will be similar to DocParagraph but will store heading level and text content.

**Severity**: `critical`
**File**: `tsdoc/src/nodes/DocHeading.ts`

## Solution

Create a new DocNode type to represent Markdown headings in the TSDoc AST. This will be similar to DocParagraph but will store heading level and text content.

## Changes

- `tsdoc/src/nodes/DocHeading.ts` (new)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

Closes #197

### DIFF
--- a/tsdoc/src/nodes/DocHeading.ts
+++ b/tsdoc/src/nodes/DocHeading.ts
@@ -1,0 +1,63 @@
+// Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
+// See LICENSE in the project root for license information.
+
+import { DocNode, type IDocNodeParameters } from './DocNode';
+import { DocNodeContainer, type IDocNodeContainerParameters } from './DocNodeContainer';
+import { DocNodeKind } from './DocNodeKind';
+
+/**
+ * Constructor parameters for {@link DocHeading}.
+ */
+export interface IDocHeadingParameters extends IDocNodeContainerParameters {
+  /**
+   * The heading level from 1 to 6, similar to Markdown headings.
+   */
+  headingLevel?: number;
+}
+
+/**
+ * Represents a heading section in the documentation, similar to Markdown headings (#, ##, etc.).
+ *
+ * @remarks
+ * This node type stores a heading level (1-6) and contains the heading content as child nodes.
+ */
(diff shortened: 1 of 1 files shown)


lukebaze:
@microsoft-github-policy-service agree

PR #485, opened 2026-09-11
Honor TokenReader subrange bounds in peek and backtrack.

### LINKED ISSUE #481: TokenReader subrange bounds enforced inconsistently: peekToken/peekPreviousTokenKind/backtrackToMarker ignore embedded window
## Summary

`TokenReader` respects its embedded subrange (`_readerStartIndex`/`_readerEndIndex`) in `peekTokenKind()`/`peekTokenAfterKind()`/`peekTokenAfterAfterKind()`, but three sibling methods ignore those bounds: `peekToken()` has no end guard at all, and `peekPreviousTokenKind()` / `backtrackToMarker()` compare against `0` / allow rewinding before the subrange start. This leaks outer tokens into embedded parses and can return `undefined` where `Token` is promised.

## Location

- File: `tsdoc/src/parser/TokenReader.ts`
- Class: `TokenReader`
- Methods:
  - `peekToken(): Token` — `return this.tokens[this._currentIndex];` with no `_readerEndIndex` check
  - `peekPreviousTokenKind()` — `if (this._currentIndex === 0)` instead of `=== this._readerStartIndex`
  - `backtrackToMarker(marker)` — checks `marker > this._currentIndex` but not `marker < this._readerStartIndex`

Contrast with the guarded siblings in the same file:

```ts
public peekTokenKind(): TokenKind {
  if (this._currentIndex >= this._readerEndIndex) {
    return TokenKind.EndOfInput;
  }
  ...
}
```

## Problem

1. **`peekToken()` past end returns `undefined`:** return type claims `Token`, but past `_readerEndIndex` the index expression yields `undefined`. Callers doing `peekToken().range` / `peekToken().kind` then throw `TypeError: Cannot read properties of undefined` instead of getting a clean `EndOfInput` signal. Current internal hot paths in `NodeParser.ts` happen to call `peekTokenKind()` first (e.g. block/inline tag `badCharacter` paths), which masks the hole, but the public API contract is broken for any external consumer or future internal use.
2. **`peekPreviousTokenKind()` leaks across embedded start:** for an embedded reader starting at index N>0, calling it at the embedded start returns `tokens[N-1].kind` (outer context) instead of `EndOfInput`. `NodeParser._parseBlockTag` and `_parseFencedCode` switch on this to decide start-of-input behavior (`AtSignInWord`, `CodeFenceOpeningIndent`); an embedded `@` or fence at the subrange start can therefore be misclassified using the token *before* the subrange.
3. **`backtrackToMarker()` can rewind before subrange start:** nothing prevents `marker < _readerStartIndex`, letting a later `readToken()` consume outer tokens that the embedded reader was explicitly scoped to exclude.

## Trigger / Reproduction

Based on static analysis (no execution performed):

- Construct `new TokenReader(parserContext, embeddedSequence)` where `embeddedSequence.startIndex > 0`, advance to the embedded start, and call `peekPreviousTokenKind()` — returns the outer predecessor kind instead of `EndOfInput`.
- Call `peekToken()` when `_currentIndex === _readerEndIndex` — returns `undefined` instead of signaling end (compare `peekTokenKind()` returning `EndOfInput` in the same state).
- Call `backtrackToMarker(outerMarker)` with `outerMarker < _readerStartIndex` on an embedded reader — accepted, subsequent reads escape the subrange.

Note: this is a static-analysis finding; I did not execute a parser fixture.

## Expected Behavior

All `TokenReader` accessors/mutators should honor the same `[ _readerStartIndex, _readerEndIndex )` window: `peekToken()` should signal end (or throw a clear parser-bug error like `readToken()` does) instead of returning `undefined`; `peekPreviousTokenKind()` should return `EndOfInput` at the subrange start; `backtrackToMarker()` should reject markers below the subrange start.

## Actual Behavior

Subrange bounds are enforced inconsistently, so embedded parses can observe outer tokens and end-of-input handling differs by method.

## Impact

- Potential spurious `AtSignInWord` / fence-indent errors for embedded constructs starting at a subrange boundary.
- `undefined`-token `TypeError`s for API consumers using `peekToken()` at end, bypassing the clean `EndOfInput` protocol every sibling method follows.

## Suggested Direction

- Mirror the existing `peekTokenKind()` guard in `peekToken()` (return the `EndOfInput` token or throw a parser-bug error — maintainer's choice, but document it), change the `peekPreviousTokenKind()` zero-check to `_readerStartIndex`, and add a `marker < _readerStartIndex` rejection in `backtrackToMarker()`. No grammar changes needed.

## Evidence

- Source via API: `tsdoc/src/parser/TokenReader.ts` shows guarded `peekTokenKind`/`peekTokenAfterKind`/`peekTokenAfterAfterKind` vs unguarded `peekToken`/`peekPreviousTokenKind`/`backtrackToMarker`; `tsdoc/src/parser/NodeParser.ts` shows `peekPreviousTokenKind` gating `AtSignInWord`/`CodeFenceOpeningIndent` and embedded `TokenReader` construction for scoped parses.
- Duplicate check: issue search for `peekToken TokenReader bounds` returns `total_count: 0`, and open issues contain no TokenReader-boundary report — no apparent duplicate. This is a non-security correctness finding, so Microsoft's private-security-disclosure requirement does not apply.

## Classification

- FACT: three `TokenReader` methods ignore the subrange window that three sibling methods enforce (verified in source via API).
- INFERENCE: embedded parses can observe out-of-range tokens; end-of-input `peekToken()` yields `undefined`.
- HYPOTHESIS: aligning all methods on the same window fixes the misclassification/TypeError risk with no grammar change.

cpruijsen:
`TokenReader.peekToken()`, `peekPreviousTokenKind()`, and `backtrackToMarker()` now honor the same `[_readerStartIndex, _readerEndIndex)` window that `peekTokenKind()` / `peekTokenAfterKind()` / `peekTokenAfterAfterKind()` already enforce. Fixes #481: an embedded reader could observe the token before its start or after its end, and `backtrackToMarker` could rewind into the outer stream. `TokenReader` is not a public export. Current `NodeParser` hot paths already call `peekTokenKind()` first, which is why existing parser snapshots did not fail.

## Decision

`peekToken()` past the window returns the stream's `EndOfInput` token (`tokens[tokens.length - 1]`); `peekPreviousTokenKind()` treats `_readerStartIndex` as start-of-input; `backtrackToMarker` throws if `marker < _readerStartIndex`.

Alternative: throw from `peekToken()` (like `readToken()`), or clamp an out-of-range marker to the window start. Open PR #482 takes the throw-from-`peekToken()` option.

`readToken()` already states that peek must always return a valid `Token`, and the other peek methods already signal `EndOfInput` at the window edge. Expired markers already throw as a parser bug. Can switch to throw-on-peek or clamp-on-backtrack.

## Test plan

- [x] New `tsdoc/src/parser/__tests__/TokenReader.test.ts` fails without the source change and passes with it (embedded predecessor leak, next-token leak, empty `TokenSequence.createEmpty`, out-of-range backtrack).
- [x] Full `@microsoft/tsdoc` Heft test run (typecheck / ESLint / API Extractor / 259 Jest tests).
- [ ] `rush change --verify` on CI (change file included).
- [ ] Optional: playground / api-demo smoke; no parser grammar change.

## Change log

Rush change file: `common/changes/@microsoft/tsdoc/fix-tokenreader-bounds_2026-09-11-13-00.json` (`patch`).

### DIFF
--- a/common/changes/@microsoft/tsdoc/fix-tokenreader-bounds_2026-09-11-13-00.json
+++ b/common/changes/@microsoft/tsdoc/fix-tokenreader-bounds_2026-09-11-13-00.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/tsdoc",
+      "comment": "Fix TokenReader peek and backtrack methods so they honor an embedded token sequence's bounds.",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@microsoft/tsdoc"
+}
--- a/tsdoc/src/parser/TokenReader.ts
+++ b/tsdoc/src/parser/TokenReader.ts
@@ -120,6 +120,10 @@ export class TokenReader {
    * consuming anything.
    */
   public peekToken(): Token {
+    if (this._currentIndex >= this._readerEndIndex) {
+      // Always return a real Token, matching peekTokenKind() === EndOfInput.
+      return this.tokens[this.tokens.length - 1];
+    }
     return this.tokens[this._currentIndex];
   }
 
@@ -180,7 +184,7 @@ export class TokenReader {
    * Returns the kind of the token immediately before the current token.
    */
   public peekPreviousTokenKind(): TokenKind {
-    if (this._currentIndex === 0) {
+    if (this._currentIndex === this._readerStartIndex) {
       return TokenKind.EndOfInput;
     }
     return this.tokens[this._currentIndex - 1].kind;
@@ -200,6 +204,10 @@ export class TokenReader {
     if (marker > this._currentIndex) {
       // If this happens, it's a parser bug
       throw new Error('The marker has expired');
+    }
+    if (marker < this._readerStartIndex) {
+      // If this happens, it's a parser bug
+      throw new Error('The marker is outside the TokenReader range');
     }
 
     this._currentIndex = marker;
--- a/tsdoc/src/parser/__tests__/TokenReader.test.ts
+++ b/tsdoc/src/parser/__tests__/TokenReader.test.ts
@@ -1,0 +1,84 @@
+// Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
+// See LICENSE in the project root for license information.
+
+import { TSDocParser } from '../TSDocParser';
+import { type Token, TokenKind } from '../Token';
+import { TokenReader } from '../TokenReader';
+import { TokenSequence } from '../TokenSequence';
+import type { ParserContext } from '../ParserContext';
+
+function parseComment(buffer: string): ParserContext {
+  const tsdocParser: TSDocParser = new TSDocParser();
+  return tsdocParser.parseString(buffer);
+}
+
+function indexOfTokenText(parserContext: ParserContext, text: string): number {
+  const tokens: ReadonlyArray<Token> = parserContext.tokens;
+  for (let i: number = 0; i < tokens.length; ++i) {
+    if (tokens[i].toString() === text) {
+      return i;
+    }
+  }
+  throw new Error('Token not found: ' + JSON.stringify(text));
+}
+
+function createEmbeddedReader(
+  parserContext: ParserContext,
+  startIndex: number,
+  endIndex: number
+): TokenReader {
+  return new TokenReader(
+    parserContext,
+    new TokenSequence({
+      parserContext: parserContext,
+      startIndex: startIndex,
+      endIndex: endIndex
+    })
+  );
+}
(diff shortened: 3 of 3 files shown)


sylvesterkaczmarek (commented):
At an embedded reader boundary, peekToken() now returns the tokenizer's global EndOfInput token. Its range is at the end of the whole comment, not readerEndIndex. NodeParser uses peekToken().range for diagnostics, so errors at an embedded boundary can point at the wrong location. Can this return an EOI token/range at the subrange end?

